PR #4, opened 2023-10-15
feat: adding waf and update tests

venkatamutyala:
chore: bump version for release 0.4.0
chore: bump verison of k3s from 1.27.2 -> 1.27.5 (matching EKS).
chore: bump version of external secrets 0.8.3 -> 0.9.5
chore: Polaris 8.0.0 -> 8.5.1
chore: kubeconform 0.6.2 -> 0.6.3
chore: kube score 1.16.1 -> 1.17.0
fix: Update polaris config. Commented out metadataAndNameMismatched as it appears to not exist at all and was probably being ignored on the older version. And added metadataAndInstanceMismatched as a warning since danger causes existing resources to fail.

### DIFF
--- a/examples/.polaris.yaml
+++ b/examples/.polaris.yaml
@@ -6,7 +6,7 @@ checks:
   pullPolicyNotAlways: ignore
   readinessProbeMissing: danger
   livenessProbeMissing: danger
-  metadataAndNameMismatched: danger
+  #metadataAndNameMismatched: danger
   pdbDisruptionsIsZero: danger
   missingPodDisruptionBudget: danger
   topologySpreadConstraint: danger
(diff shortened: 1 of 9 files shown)


examples/.polaris.yaml
```diff
@@ -6,7 +6,7 @@ checks:
   pullPolicyNotAlways: ignore
   readinessProbeMissing: danger
   livenessProbeMissing: danger
-  metadataAndNameMismatched: danger
```
venkatamutyala, line 9: @debu99 i have a feeling on the older version of Polaris `metadataAndNameMismatched` didn't exist and was not being caught. On the newer version it seems to error out when I keep this in the config. So I switched to: `metadataAndInstanceMismatched` and that throws a danger /error so i went ahead and added `metadataAndInstanceMismatched` as a warning. I checked the Polaris docs on the fairwinds site and it seems like `metadataAndInstanceMismatched` is the correct key to use not `metadataAndNameMismatched`. Curious to know your thoughts and whether we should roll back what I've done here.

fernandoataoldotcom (approved):

PR #116, opened 2026-05-19
fix(keda): render ScaledObject nested spec fields as YAML

yesterdaysrebel:
Render non-trigger keda.scaledObject.spec values as proper YAML for maps, lists, and scalars (including YAML-special strings), instead of Go-style string interpolation. Add regression tests for nested spec fields, scalars, scaleTargetRef, and trigger types.

Removes unrelated ExternalSecret injectIntoPod, legacy RabbitMQ ScaledObject template, and chart version bumps that were combined into this branch.

Releated issues - 

https://github.com/internal-GlueOps/issues/issues/308

https://github.com/internal-GlueOps/issues/issues/304

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ A Helm chart template for applications
 | job.jobs | string | `nil` | List of jobs to run |
 | keda | object | `{"enabled":false,"scaledObject":{"enabled":false,"spec":{"triggers":[],"triggersMap":{}}},"triggerAuthentication":null}` | Keda configuration |
 | keda.enabled | bool | `false` | Whether to enable Keda |
-| keda.scaledObject | object | `{"enabled":false,"spec":{"triggers":[],"triggersMap":{}}}` | ScaledObject's spec |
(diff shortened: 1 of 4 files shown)


venkatamutyala (changes requested):
**`values.yaml:1035`** — Now that nested fields render correctly, can we expand the commented example here to include `advanced.scalingModifiers` (GA since KEDA 2.15) so the unlock shows up in helm-docs? helm-docs only walks parsed YAML keys, so leaving the block commented keeps it from becoming a live default for every tenant — but a `# --` annotation on the parent keys would surface them in the generated README. A one-line `# --` caveat that string values matching numeric/boolean patterns (e.g. `pollingInterval: "30"`) get rendered as their native type would also help avoid surprises.

Otherwise the fix and tests look solid.

yesterdaysrebel:
@venkatamutyala - I have added the comment as requested. This PR is ready for review.

venkatamutyala (approved):

PR #117, opened 2026-05-20
feat: global imagePullSecrets inheritance for all pod templates (#326)

yesterdaysrebel:
## Summary

Fixes [internal-GlueOps/issues#326](https://github.com/internal-GlueOps/issues/issues/326) by adding global registry pull-secret inheritance for **Deployment**, **StatefulSet**, **Job**, and **CronJob** pod templates.

- Set `image.pullSecrets` once in common values instead of repeating `cronJob.jobs.<name>.imagePullSecrets` on every job.
- Set `cronJob.imagePullSecrets` / `job.imagePullSecrets` as workload-wide defaults for all jobs in that workload.
- Add `cronJob.jobs.<name>.enabled: false` / `job.jobs.<name>.enabled: false` to opt out of jobs defined in shared values.
- Bump chart version **0.13.0 → 0.14.0**.

## How it works

New helpers:

| Helper | Purpose |
|--------|---------|
| `chart.imagePullSecretName` | Resolves the secret name with documented precedence |
| `chart.jobPodGlobals` | Merges the workload-global `imagePullSecrets` (only) from `cronJob`/`job` into each job's pod context |
| `chart.jobEntryEnabled` | Validates `enabled` is boolean when set |
| `chart.assertPullSecretKind` | Rejects set-but-non-string values, naming the offending values path |

**Precedence (highest wins):**

1. `cronJob.jobs.<name>.imagePullSecrets` / `job.jobs.<name>.imagePullSecrets`
2. Workload `imagePullSecrets` (`deployment`, `statefulSet`, `cronJob`, `job`)
3. `image.pullSecrets`

**One rule at every level.** A level is *set* when the key is present and non-null; the first set level wins, **even when its value is empty**. The same literal means the same thing everywhere:

| Value | Meaning |
|-------|---------|
| `"my-cred"` | Use this secret |
| `""` | Opt out — render no pull secret, ignore every level below |
| `null` / absent | Unset — inherit from the next level down |

`""` is the **only** opt-out. Other falsey values (`false`, `0`, `[]`) are rejected as type errors rather than rendered as a literal secret name (`- name: "false"`).

> **Naming note:** the top-level key is `image.pullSecrets`, but every workload/per-job override key is `imagePullSecrets` (not `pullSecrets`). Using `pullSecrets` on a workload is a silent no-op.

## Example

**Common values file** — carries the shared secret only. Note there is deliberately **no `cronJob.jobs.*` here**: declaring a job in a common file is what created a phantom CronJob in every app, which is the bug this PR fixes.

```yaml
image:
  pullSecrets: my-registry-cred

# Optional: a default for all CronJob pods, without declaring any job.
cronJob:
  imagePullSecrets: my-registry-cred
```

**App values file** — the app declares its own jobs, with any names it likes. They all inherit the secret; no per-job `imagePullSecrets` needed, and no shared job-name convention.

```yaml
cronJob:
  enabled: true
  jobs:
    nightly-backup:
      schedule: "0 0 * * *"
      image: my/image:tag
    hourly-sync:
      schedule: "0 * * * *"
      image: my/other-image:tag
```

**Opting out.** Any level can escape a secret set in a common values file with `""`:

```yaml
deployment:
  imagePullSecrets: ""      # this Deployment pulls without a secret

cronJob:
  jobs:
    public-image-job:
      imagePullSecrets: ""  # just this job
```

`enabled: false` is the equivalent escape hatch for a whole job a common values file declares:

```yaml
cronJob:
  jobs:
    job-from-common-i-dont-want:
      enabled: false
```

## Upgrade notes (0.13.0 → 0.14.0)

Two inputs that were silently **inert** on 0.13.0 become **live** on 0.14.0, and one class of malformed input now fails loudly. No tenant in the current fleet sets any of them.

- `image.pullSecrets` now injects `imagePullSecrets` into every pod template. Previously ignored entirely.
- `cronJob.jobs.<name>.enabled: false` / `job.jobs.<name>.enabled: false` now **removes** that CronJob/Job. Under ArgoCD `prune: true` this prunes a live object. Previously the key was ignored and the job rendered anyway.
- A non-string `*.imagePullSecrets` (int, list, map, whitespace-only, or a name containing whitespace) now **fails the render** with the offending values path. Previously it rendered a malformed PodSpec (e.g. `- name: [a b]`).

**Workload `imagePullSecrets` behavior is unchanged from 0.13.0.** Verified against `origin/main` by rendering both:

| Input | 0.13.0 | 0.14.0 |
|-------|--------|--------|
| `deployment.imagePullSecrets: cred` | `- name: cred` | `- name: "cred"` |
| `deployment.imagePullSecrets: ""` | no pull secret | no pull secret |
| `image.pullSecrets: G` | no pull secret (inert) | `- name: "G"` (new) |

The only rendering difference on the existing key is that the name is now quoted — see the `| quote` note below.

Adopters of the new keys should pin `>= 0.14.0`: rolling back to 0.13.0 makes them inert again.

## Review feedback addressed

**First round**

- **B1 — version/rebase:** rebased onto current `main` (was forked from #115 and reverted #116's KEDA rendering); bumped to `0.14.0`. `git diff main` now contains only the imagePullSecrets work — no KEDA changes.
- **B2 — key naming:** the helper reads `image.pullSecrets`, the customer-requested key.
- **M1 — nil panic:** replaced `ternary` (which eagerly evaluated both `cronJob.jobs` and `job.jobs`) with `if/else` on `resourceType`, and null-guarded line 1 of `cronjob.yaml`/`job.yaml`. `cronJob: null` / `job: null` no longer panic.
- **M2 — scope creep:** `chart.jobPodGlobals` was allowlisting ~35 pod fields to inherit from `cronJob`/`job` globals. Narrowed to `imagePullSecrets` only, so no other pod-spec fields silently change on upgrade.

**Second round** (all five reproduced locally before fixing; `5fc4c61`)

- **Alias cut.** `image.imagePullSecrets` is gone. Two spellings for one thing bought a precedence rule ("honored only when `pullSecrets` is unset or empty") that nobody would remember and that had already generated three review comments just keeping its own docs self-consistent — a bad trade for a key that tags as permanent API. `image.pullSecrets` is the only top-level key; a test asserts the old spelling is now inert.
- **`""` now means one thing everywhere.** Previously workload-level `""` was falsy and *inherited*, while per-job `""` *opted out* — the same literal with opposite meanings by level, leaving no way to opt a Deployment out of a global set in a common values file. Selection is now `hasKey` + non-null at every level. This also required flipping the workload defaults in `values.yaml` from `""` to `null`, since with a `""` default the new semantics would have opted the entire fleet out. As the table above shows, the uniform rule is the one that **preserves** 0.13.0 behavior; the earlier draft of this PR was the silent change.
- **Cross-job leakage was untested.** `index $jobsMap .name` sits inside a `range`, so a constant key gave every job the first job's secret with 318/318 still green. Added 3-job tests (cronJob and job) with disagreeing secrets, where the third job is deliberately secret-less so the fall-through is pinned too, plus a scoped `""` opt-out test.
- **Name guards were untested.** Removing the whitespace guard left 318/318 green and `image.pullSecrets: "my secret"` rendered `- name: "my secret"`. Kubernetes does not validate pull secret names at the API server, so this render-time check is the only thing that catches a typo — that reasoning is now recorded in the test block. Coverage added for internal spaces, tabs, whitespace-only, trimming, and the per-job path.
- **`| quote` is load-bearing, now documented.** Helm's loader is YAML 1.1, so an unquoted `y`/`no`/`on`/`off` secret name parses as a bool. helm-unittest cannot catch the quote's removal — it reads the rendered doc with a YAML 1.2 parser, which normalizes the difference away — so a green suite is not permission to simplify it away.

Each fix was verified by mutation rather than by a green suite: the constant-job-key mutant now fails 3 tests, the whitespace guard 4, the empty guard 2.

## Test plan

- [x] `helm lint .`
- [x] `helm unittest .` (324 passed)
- [x] Every `examples/` values file renders
- [x] Mutation-tested: each new guard has a test that fails when the guard is removed
- [x] 0.13.0 vs 0.14.0 render diff for workload `imagePullSecrets` (table above)
- [x] Rebased onto `main`; KEDA nested-spec rendering (#116) intact and tested
- [ ] Consumer apps can adopt `image.pullSecrets` in common values after chart bump
- [ ] Magic Memories / customer cronjobs validated in nonprod after chart dependency update

## Consumer impact

- **No change required** for apps already setting per-job or workload `imagePullSecrets` — those render as they did on 0.13.0.
- **No KEDA impact** — after the rebase this branch no longer reverts #116.
- **Optional:** move the shared secret to `image.pullSecrets` or `cronJob.imagePullSecrets` and remove duplicated per-job entries after upgrading to chart `0.14.0`.


venkatamutyala (changes requested):
**Requesting changes — verified locally against this branch. Three items, then good to go.**

Nice work, and thanks for tackling the whole of #326 — the `enabled: false` opt-out and pod-template defaults are both genuinely in-scope per the customer's report, not just the pull secret.

**1. Per-job pod-template overrides lose to the global default (must-fix).**
A tenant who sets a per-job field *and* a workload-level default gets the global, not theirs — the opposite of the documented "per-job overrides global." E.g. `cronJob.tolerations` + `cronJob.jobs.<name>.tolerations` → the global wins. Cause: in `templates/cronjob.yaml:53` / `job.yaml:35`, `merge (dict …) $podGlobals $job` — Sprig `merge` is non-overwriting and the first source populates first, so `$podGlobals` beats `$job`. This affects all inherited pod fields (`tolerations`, `affinity`, `nodeSelector`, …); `imagePullSecrets` is the one exception, since its helper reads `Values` directly.
Fix: swap the order → `merge (dict …) $job $podGlobals`, plus a regression test for a non-`imagePullSecrets` field (e.g. per-job `tolerations` over global).

**2. Whitespace-only secret name renders nothing, silently (must-fix).**
`image.imagePullSecrets: "   "` is truthy, then trims to empty → no `imagePullSecrets` block, no error. (Interior whitespace like `"a b"` correctly `fail`s — only the all-whitespace case slips through.) Either `fail` on it like the other validation paths, or normalize-then-check so it behaves as unset.

**3. Drop the `image.pullSecrets` alias — keep just `image.imagePullSecrets`.**
Both names are net-new, so no compat reason to ship two. We weighed the single name from three angles — product (consistency/DX), platform (multi-tenant base+overlay safety), and the end-user/consumer perspective — and the deciding factor was the consumer view: a chart user manages one cred across `deployment.imagePullSecrets`, `cronJob.imagePullSecrets`, per-job, and the image level, so keeping `image.imagePullSecrets` lets them grep/copy-paste one identical key across every scope. The odd-one-out name is the real footgun; the mild stutter under `image:` was judged an acceptable cost. (It also matches the K8s field name.) Cleanup spans the dual-name precedence tier + the "pullSecrets beats imagePullSecrets" tiebreak in `_helpers.tpl`, the `image.pullSecrets` references in `values.yaml`, and the alias test.

Versioning: all three are pre-release adjustments to net-new code — still a clean MINOR (0.13.0), no major bump needed.

_Non-blocking heads-up:_ `| quote` at `_podTemplate.tpl:81` renders `- name: x` as `- name: "x"`, so existing pull-secret consumers will see a one-time cosmetic diff on their next ArgoCD/Flux sync — worth a release-note line.

Everything else — precedence, opt-out, type validation, the unittest suite + `helm lint` — checks out. 👍

yesterdaysrebel:
> **Requesting changes — verified locally against this branch. Three items, then good to go.**
> 
> Nice work, and thanks for tackling the whole of #326 — the `enabled: false` opt-out and pod-template defaults are both genuinely in-scope per the customer's report, not just the pull secret.
> 
> **1. Per-job pod-template overrides lose to the global default (must-fix).** A tenant who sets a per-job field _and_ a workload-level default gets the global, not theirs — the opposite of the documented "per-job overrides global." E.g. `cronJob.tolerations` + `cronJob.jobs.<name>.tolerations` → the global wins. Cause: in `templates/cronjob.yaml:53` / `job.yaml:35`, `merge (dict …) $podGlobals $job` — Sprig `merge` is non-overwriting and the first source populates first, so `$podGlobals` beats `$job`. This affects all inherited pod fields (`tolerations`, `affinity`, `nodeSelector`, …); `imagePullSecrets` is the one exception, since its helper reads `Values` directly. Fix: swap the order → `merge (dict …) $job $podGlobals`, plus a regression test for a non-`imagePullSecrets` field (e.g. per-job `tolerations` over global).
> 
> **2. Whitespace-only secret name renders nothing, silently (must-fix).** `image.imagePullSecrets: " "` is truthy, then trims to empty → no `imagePullSecrets` block, no error. (Interior whitespace like `"a b"` correctly `fail`s — only the all-whitespace case slips through.) Either `fail` on it like the other validation paths, or normalize-then-check so it behaves as unset.
> 
> **3. Drop the `image.pullSecrets` alias — keep just `image.imagePullSecrets`.** Both names are net-new, so no compat reason to ship two. We weighed the single name from three angles — product (consistency/DX), platform (multi-tenant base+overlay safety), and the end-user/consumer perspective — and the deciding factor was the consumer view: a chart user manages one cred across `deployment.imagePullSecrets`, `cronJob.imagePullSecrets`, per-job, and the image level, so keeping `image.imagePullSecrets` lets them grep/copy-paste one identical key across every scope. The odd-one-out name is the real footgun; the mild stutter under `image:` was judged an acceptable cost. (It also matches the K8s field name.) Cleanup spans the dual-name precedence tier + the "pullSecrets beats imagePullSecrets" tiebreak in `_helpers.tpl`, the `image.pullSecrets` references in `values.yaml`, and the alias test.
> 
> Versioning: all three are pre-release adjustments to net-new code — still a clean MINOR (0.13.0), no major bump needed.
> 
> _Non-blocking heads-up:_ `| quote` at `_podTemplate.tpl:81` renders `- name: x` as `- name: "x"`, so existing pull-secret consumers will see a one-time cosmetic diff on their next ArgoCD/Flux sync — worth a release-note line.
> 
> Everything else — precedence, opt-out, type validation, the unittest suite + `helm lint` — checks out. 👍

@venkatamutyala - All review comments are fixed now. Request you to review this again.

Chart.yaml
```diff
@@ -1,4 +1,4 @@
 name: app
 apiVersion: v2
 description: A Helm chart template for applications
-version: 0.12.0
+version: 0.13.0
```
venkatamutyala: **Blocker (B1): `0.13.0` is already a released, tagged version with different content.** `v0.13.0` is tagged and is main's current tip (#118's bump), and that release does not contain this feature — `git diff v0.13.0 <this-branch> -- templates/_helpers.tpl` differs. This branch forked from `0.12.0` and re-bumps to the same `0.13.0`, so merging republishes different chart contents under an already-published version (ArgoCD consumers pinned to `0.13.0` would split across two digests with no OutOfSync signal). Note a `git merge main` auto-resolves this file cleanly since both sides say `0.13.0`, so nothing catches it. Please rebase onto current main and bump to `0.14.0`.

templates/_helpers.tpl
```diff
@@ -162,6 +162,71 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+{{- $explicitPerJob = true -}}
+{{- $secret = .imagePullSecrets -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- if not $explicitPerJob -}}
+{{- if .imagePullSecrets -}}
+{{- $secret = .imagePullSecrets -}}
+{{- else if and (kindIs "map" .Root.Values.image) .Root.Values.image.imagePullSecrets -}}
+{{- $secret = .Root.Values.image.imagePullSecrets -}}
```
venkatamutyala: **Blocker (B2): this is the only place the top-level pull secret is read, and it reads `image.imagePullSecrets` — but the PR body example (`image.pullSecrets`), the precedence table (steps 3–4), and 4 `values.yaml` comments all tell users to set `image.pullSecrets`, which is silently ignored.**

```
--set image.pullSecrets=cred        → no imagePullSecrets rendered (SILENT no-op)
--set image.imagePullSecrets=cred   → imagePullSecrets: [- name: "cred"]
```
John requested `image.pullSecrets` (with `image.imagePullSecrets` as an acceptable alternative). Pick one name and make code + docs + PR body agree; adding an `image.pullSecrets` read here matches the customer request. Add a test on the documented key so it can't regress.

values.yaml
```diff
@@ -83,6 +87,8 @@ commonLabels:
 deployment:
   # -- Whether to create a Deployment
   enabled: false
+  # -- Registry pull secret for this Deployment (overrides `image.pullSecrets`). Empty string inherits `image.pullSecrets`.
```
venkatamutyala: B2: `image.pullSecrets` is not a key the chart reads — the helper reads `image.imagePullSecrets`. This prose (and lines 356, 630, 739) sends users to a dead key. Correct to `image.imagePullSecrets` (or add the `pullSecrets` alias in the helper).

values.yaml
```diff
@@ -618,6 +626,10 @@ cronJob:
   # -- Whether to create a CronJob
   enabled: false
 
+  # -- Registry pull secret for all CronJob pods (per-job `imagePullSecrets` overrides).
+  # Falls back to top-level `image.pullSecrets` when unset.
```
venkatamutyala: B2 (same as line 90): `image.pullSecrets` → `image.imagePullSecrets`. The `cronJob`/`job` fallback comments both point at the non-existent key.

templates/_helpers.tpl
```diff
@@ -162,6 +162,71 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+  1. cronJob.jobs.<name>.imagePullSecrets / job.jobs.<name>.imagePullSecrets when the key is present (only this path: empty string opts out)
+  2. .imagePullSecrets on the merged workload context when truthy (deployment/statefulSet empty string is falsy and inherits step 3)
+  3. .Root.Values.image.imagePullSecrets (when .Values.image is a map)
+Per-job lookup uses Values.jobs directly because shallow merge does not reliably override globals.
+*/}}
+{{- define "chart.imagePullSecretName" -}}
+{{- $secret := "" -}}
+{{- $explicitPerJob := false -}}
+{{- if and .name (or (eq .resourceType "cronJob") (eq .resourceType "job")) -}}
+{{- $jobsMap := ternary .Root.Values.cronJob.jobs .Root.Values.job.jobs (eq .resourceType "cronJob") -}}
```
venkatamutyala: **Major (M1): nil-panic.** `ternary` evaluates *both* arguments regardless of the condition, so this dereferences both `.cronJob.jobs` and `.job.jobs`. A consumer who disables one path by setting `cronJob: null` (or `job: null`) crashes the render of the other:
```
helm template ... --set job.enabled=true --set job.jobs.run.image=nginx --set cronJob=null
Error: ... <.Root.Values.cronJob.jobs>: nil pointer evaluating interface {}.jobs
```
Use `{{- if eq .resourceType "cronJob" }}...{{- else }}...{{- end }}` so only the relevant map is referenced.

templates/_helpers.tpl
```diff
@@ -162,6 +162,71 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
 
 {{- end -}}
 
+{{/*
+Pod-template fields allowed to inherit from cronJob/job global config (excludes labels, annotations, JobSpec, schedule).
+*/}}
+{{- define "chart.jobPodGlobals" -}}
+{{- toYaml (pick . "imagePullSecrets" "tolerations" "affinity" "nodeSelector" "topologySpreadConstraints" "topologySpreadConstraintsMaxSkew" "image" "imagePullPolicy" "command" "args" "resources" "securityContext" "containerSecurityContext" "envMap" "envSecrets" "envFrom" "envVariables" "envConfigMaps" "volumes" "volumeMounts" "initContainers" "sidecar" "serviceAccount" "enableServiceLinks" "hostNetwork" "hostAliases" "dnsPolicy" "dnsConfig" "terminationGracePeriodSeconds" "priorityClassName" "shareProcessNamespace" "restartPolicy" "matchLabels" "startupProbe" "livenessProbe" "readinessProbe") -}}
```
venkatamutyala: **Major (M2): scope creep + latent backwards-compat risk.** John asked for one field (`imagePullSecrets`), but this allowlists ~35 pod fields to inherit from `cronJob`/`job` globals — previously ignored. Default renders are unchanged, but: (1) any tenant with a stray global `cronJob.image`/`resources`/etc. gets silently changed pods on upgrade; (2) a malformed global `envMap` that was harmlessly ignored now renders invalid YAML → hard-fails the render; (3) `restartPolicy` here lets a global `Always` produce an API-invalid Job pod; (4) the hand-kept 35-key list will drift from `_podTemplate.tpl`. Recommend narrowing to `imagePullSecrets` (+ anything deliberately in scope), or splitting the broad passthrough into its own versioned, documented change.

templates/cronjob.yaml
```diff
@@ -48,8 +49,10 @@ spec:
   jobTemplate:
     spec:
       {{- include "chart.jobSpec" (dict "global" $.Values.cronJob "job" $job) | indent 6 }}
-      {{- $context :=  merge (dict "resourceType" "cronJob" "name" $name "Chart" $.Chart "Root" $newContext) $job }}
+      {{- $podGlobals := fromYaml (include "chart.jobPodGlobals" (omit $.Values.cronJob "jobs" "enabled")) | default dict -}}
```
venkatamutyala, line 52: This `merge ... $podGlobals` is where the M2 behavior change activates: global `cronJob.*` pod fields (image, resources, envMap, sidecar, securityContext, …) now flow into every job pod, whereas on `main` they were ignored. Same in `job.yaml:34`. Fine if intentional and scoped/versioned as such — flagging so it's a deliberate decision, not a side effect of the pull-secret work.

venkatamutyala (changes requested):
Thanks for tackling this — the core feature works and it solves **both** halves of John's ask: a single global secret is inherited by Deployment/StatefulSet/Job/CronJob, and `cronJob.jobs.<name>.enabled: false` cleanly fixes his "can't opt a shared-values job out" blocker (nicely, the boolean is enforced so the YAML `"false"`-string footgun hard-fails). `helm lint` is clean and 262 unit tests pass.

Requesting changes on **2 blockers + a render-crash + a scope concern**. All points below were reproduced with `helm template` on this branch.

### 🔴 Blockers
**B1 — version `0.13.0` is already released with different content.** `v0.13.0` is tagged and is main's current tip (#118's bump); that release does not contain this feature. This branch forked from `0.12.0` and re-bumps to `0.13.0`, so it would publish different chart contents under an already-published version. A `git merge main` auto-resolves cleanly (both sides say `0.13.0`), so nothing flags it. Please rebase onto current main and bump to **`0.14.0`**.

**B2 — the documented / customer-requested key `image.pullSecrets` is a silent no-op.** The helper only reads `image.imagePullSecrets`. `--set image.pullSecrets=cred` → no `imagePullSecrets` rendered; `--set image.imagePullSecrets=cred` → works. The PR body example, the precedence table (steps 3–4), and 4 `values.yaml` comments all point at the dead key — a user copying the example gets fleet-wide `ImagePullBackOff` with no build-time error. Pick one name (adding an `image.pullSecrets` read matches John's request) and make code + all docs + PR body agree, with a test on the documented key.

### 🟠 Major
**M1 — `ternary` nil-panic** (`_helpers.tpl:200`): setting `cronJob: null` or `job: null` to disable one path crashes the render of the other (`nil pointer evaluating interface {}.jobs`), because `ternary` evaluates both branches.

**M2 — scope creep beyond `imagePullSecrets`:** `chart.jobPodGlobals` allowlists ~35 pod fields that now inherit from `cronJob`/`job` globals (previously ignored). Default renders are unchanged, but any tenant with a stray global `cronJob.image`/`resources`/etc. gets silently changed pods on upgrade; a malformed global `envMap` now hard-fails the render; and a global `restartPolicy: Always` yields API-invalid Job pods. Recommend narrowing to `imagePullSecrets` (+ anything deliberately in scope) or splitting the broad passthrough into its own versioned change.

### 🟡 Minor
- `""` means opposite things by depth: global `imagePullSecrets: ""` inherits `image.*`, per-job `""` opts out. Setting the global `""` to disable secrets injects `image.imagePullSecrets` anyway.
- Test gaps: no test for the documented `image.pullSecrets` key; deployment/statefulSet `""`-inherit contract untested; whitespace-only & embedded-whitespace validation branches untested. (PR body says 265 tests; actual is 262.)
- Consider a `values.schema.json` with `additionalProperties: false` on `image` — it would turn the whole unknown-key class (incl. B2) into an immediate lint error. Drop the manual README badge edit; helm-docs regenerates it.

Minimum to merge: rebase + bump to `0.14.0` (B1); reconcile the key name across code/docs/PR body and test it (B2); fix the nil-panic (M1); decide whether the 35-field passthrough is in scope (M2).

yesterdaysrebel:
Thanks for the review @venkatamutyala — all four addressed and force-pushed (rebased onto current `main`):

- **B1** — Rebased onto `main` (branch had forked from #115 and was reverting #116's KEDA rendering + deleting its tests). Bumped to **0.14.0**. `git diff main` now contains only the imagePullSecrets work; the restored #116 KEDA nested-spec tests pass.
- **B2** — Helper now reads `image.pullSecrets` (primary) with `image.imagePullSecrets` as an alias. Code + `values.yaml` + `.ai/patterns.md` + PR body now agree; added tests on the documented key.
- **M1** — Replaced `ternary` with `if/else` on `resourceType`, and null-guarded line 1 of `cronjob.yaml`/`job.yaml`. `--set cronJob=null` / `job=null` now render instead of panicking; regression tests added (and verified they fail if the guard is reverted).
- **M2** — Narrowed `chart.jobPodGlobals` to `imagePullSecrets` only. Added a guard test asserting a global `cronJob.nodeSelector` does **not** leak into job pods.

Also caught during a re-review: a per-job `imagePullSecrets: null` was rendering an invalid `name: ""`; it now inherits like an unset value.

`helm lint` clean, `helm unittest` **292 passed**. Default/representative renders are byte-identical to `main` except the chart-version label.

templates/_helpers.tpl
```diff
@@ -162,6 +162,115 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+{{- else -}}
+{{- $jobsMap = .Root.Values.job.jobs -}}
+{{- end -}}
+{{- with index $jobsMap .name -}}
+{{- if and (hasKey . "imagePullSecrets") (not (kindIs "invalid" .imagePullSecrets)) -}}
+{{- $explicitPerJob = true -}}
+{{- $secret = .imagePullSecrets -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `23366cf`.

Per-job type errors were falling through to the generic tail validation, so the message named no path. The per-job value is now kind-checked at its selection point, and a `$source` path is threaded through the remaining (empty / whitespace) checks too, so every failure names the offending key:

```
cronJob.jobs.nightly-backup.imagePullSecrets must be a string secret name, got bool (value: false). Use a secret name, or "" to opt this job out of pull secrets.
cronJob.jobs.nightly-backup.imagePullSecrets must not contain whitespace, got "a b"
cronJob.jobs.nightly-backup.imagePullSecrets must not be empty or whitespace only
job.jobs.run-once.imagePullSecrets must be a string secret name, got slice (value: [])
```

Regression tests added pinning the path-qualified messages so a refactor can't silently drop the job name.

templates/_helpers.tpl
```diff
@@ -162,6 +162,115 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+legitimate falsey inherit/opt-out signal whose meaning depends on the level.
+Inputs: .value, .key (the values path, used in the error message)
+*/}}
+{{- define "chart.assertPullSecretKind" -}}
+{{- if not (kindIs "invalid" .value) -}}
+{{- if not (kindIs "string" .value) -}}
+{{- fail (printf "%s must be a string secret name, got %s (value: %v). Use a secret name, or \"\" to opt out." .key (kindOf .value) .value) -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid, and a good catch — that wording was a regression I introduced in the previous commit. Fixed in `23366cf`.

`""` genuinely means different things by level, so a single hint string could never be right. The hint is now supplied per call site:

| level | `""` means | hint |
|---|---|---|
| `image.pullSecrets` / `image.imagePullSecrets` | unset | `remove the key to render no pull secret` |
| `<workload>.imagePullSecrets` | inherit | `"" to inherit image.pullSecrets` |
| `<workload>.jobs.<name>.imagePullSecrets` | opt out | `"" to opt this job out of pull secrets` |

Verified:
```
image.pullSecrets must be a string secret name, got bool (value: false). Use a secret name, or remove the key to render no pull secret.
deployment.imagePullSecrets must be a string secret name, got bool (value: false). Use a secret name, or "" to inherit image.pullSecrets.
```

templates/_helpers.tpl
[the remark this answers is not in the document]
yesterdaysrebel: Same finding as the earlier comment on this line — fixed in `23366cf`. Per-job values are now kind-checked at their selection point and every failure names the full path, e.g. `cronJob.jobs.nightly-backup.imagePullSecrets must be a string secret name, got bool (value: false)`. Regression tests added.

templates/_helpers.tpl
```diff
@@ -162,6 +162,124 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+{{- else if kindIs "map" .Root.Values.image -}}
+{{- if .Root.Values.image.pullSecrets -}}
+{{- $secret = .Root.Values.image.pullSecrets -}}
+{{- $source = "image.pullSecrets" -}}
+{{- else if .Root.Values.image.imagePullSecrets -}}
+{{- $secret = .Root.Values.image.imagePullSecrets -}}
+{{- $source = "image.imagePullSecrets" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
yesterdaysrebel: Half-valid — the observation is right, the suggested remedy would break the chart. Doc fixed in `990a489`; code deliberately unchanged.

**You're right that** `image.pullSecrets: ""` does not block the alias:
```
image: {pullSecrets: "", imagePullSecrets: alias-cred}  ->  - name: "alias-cred"
```

**But switching top-level selection to `hasKey` would kill the alias outright.** `pullSecrets` now carries a `""` default in values.yaml (promoted from a comment so helm-docs publishes it), so `hasKey .Values.image "pullSecrets"` is **always true** — the alias branch would become unreachable in every case, and its tests would fail.

The real defect was the **documentation**. At the top level `""` genuinely *is* unset — the chart's own error text already says `remove the key to render no pull secret`. values.yaml said the alias applies "only when `pullSecrets` is unset", which was inaccurate. Now corrected to **"unset or empty"**, with an explicit note that a top-level `""` does not block the alias. Two tests added pinning the contract (alias honored when `pullSecrets: ""`; no secret when both are empty).

Per-job `""` remains a true opt-out — that path uses `hasKey` + non-null precisely so it can distinguish explicit-empty from unset.

templates/_helpers.tpl
```diff
@@ -162,6 +162,124 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+Whether a cronJob.jobs.<name> or job.jobs.<name> entry should render. .enabled must be boolean when set.
+*/}}
+{{- define "chart.jobEntryEnabled" -}}
+{{- $job := .job -}}
+{{- $name := .name -}}
+{{- if hasKey $job "enabled" -}}
+{{- if not (kindIs "bool" $job.enabled) -}}
+{{- fail (printf "jobs.%s.enabled must be a boolean, got %s (value: %v). Use true or false, not a string." $name (kindOf $job.enabled) $job.enabled) -}}
+{{- end -}}
+{{- ternary "true" "false" $job.enabled -}}
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `990a489`. `chart.jobEntryEnabled` now takes `resourceType` (passed from `cronjob.yaml:3` / `job.yaml:3`) and reports the real values path:
```
cronJob.jobs.backup.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
job.jobs.run-once.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
```
Now consistent with the path-qualified pull-secret errors. Also dropped the trailing "not a string" clause, which was wrong for non-string non-bool inputs. Existing tests retargeted to the qualified message + new coverage added.

templates/_helpers.tpl
```diff
@@ -162,6 +162,127 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+{{- $name := .name -}}
+{{- if and (hasKey $job "enabled") (not (kindIs "invalid" $job.enabled)) -}}
+{{- if not (kindIs "bool" $job.enabled) -}}
+{{- fail (printf "jobs.%s.enabled must be a boolean, got %s (value: %v). Use true or false (unquoted)." $name (kindOf $job.enabled) $job.enabled) -}}
+{{- end -}}
+{{- ternary "true" "false" $job.enabled -}}
+{{- else -}}
+true
+{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `990a489`. `chart.jobEntryEnabled` now takes `resourceType` (passed from `cronjob.yaml:3` / `job.yaml:3`) and reports the real values path:
```
cronJob.jobs.backup.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
job.jobs.run-once.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
```
Now consistent with the path-qualified pull-secret errors. Also dropped the trailing "not a string" clause, which was wrong for non-string non-bool inputs. Existing tests retargeted to the qualified message + new coverage added.

templates/job.yaml
```diff
@@ -1,5 +1,6 @@
-{{ if .Values.job.enabled }}
+{{ if and .Values.job .Values.job.enabled }}
 {{- range $name, $job := .Values.job.jobs }}
+{{- if eq (include "chart.jobEntryEnabled" (dict "job" $job "name" $name)) "true" }}
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `990a489`. `chart.jobEntryEnabled` now takes `resourceType` (passed from `cronjob.yaml:3` / `job.yaml:3`) and reports the real values path:
```
cronJob.jobs.backup.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
job.jobs.run-once.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
```
Now consistent with the path-qualified pull-secret errors. Also dropped the trailing "not a string" clause, which was wrong for non-string non-bool inputs. Existing tests retargeted to the qualified message + new coverage added.

templates/cronjob.yaml
```diff
@@ -1,5 +1,6 @@
-{{ if .Values.cronJob.enabled }}
+{{ if and .Values.cronJob .Values.cronJob.enabled }}
 {{- range $name, $job := .Values.cronJob.jobs }}
+{{- if eq (include "chart.jobEntryEnabled" (dict "job" $job "name" $name)) "true" }}
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `990a489`. `chart.jobEntryEnabled` now takes `resourceType` (passed from `cronjob.yaml:3` / `job.yaml:3`) and reports the real values path:
```
cronJob.jobs.backup.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
job.jobs.run-once.enabled must be a boolean, got string (value: false). Use true or false (unquoted).
```
Now consistent with the path-qualified pull-secret errors. Also dropped the trailing "not a string" clause, which was wrong for non-string non-bool inputs. Existing tests retargeted to the qualified message + new coverage added.

values.yaml
```diff
@@ -67,6 +67,15 @@ image:
...
   port: 8080
+  # -- Registry pull secret name inherited by all pod templates (Deployment, StatefulSet, Job, CronJob).
+  # Resolution order (highest first): per-job `imagePullSecrets` -> workload `imagePullSecrets`
+  #   -> `image.pullSecrets` -> `image.imagePullSecrets` (alias, used only when `pullSecrets` is unset).
+  # NOTE: the top-level key here is `pullSecrets`, but every workload/per-job override key is
+  #   `imagePullSecrets` (not `pullSecrets`). Using `pullSecrets` on a workload is a silent no-op.
+  # Only cronJob.jobs.<name>.imagePullSecrets / job.jobs.<name>.imagePullSecrets: "" opts out (no pull
+  #   secret); an empty string elsewhere inherits, and a `null` per-job value behaves as unset.
+  # Example: pullSecrets: my-registry-cred
+  pullSecrets: ""
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `990a489`. `image.imagePullSecrets` is now a documented key with a `""` default, so helm-docs publishes it and the alias is discoverable rather than existing only in the helper and tests. Verified render-neutral across every `examples/*/values.yaml` and `examples/testcases/*.yaml`. It's documented as an alias with a pointer to prefer `pullSecrets`.

.ai/patterns.md
```diff
@@ -3,6 +3,9 @@
 ## Shared Helpers (`_helpers.tpl`)
 
 - **`chart.jobSpec`** — Data-driven helper rendering JobSpec fields for both Job and CronJob. Uses `$intFields`, `$stringFields`, and `$objectFields` lists. To add a new K8s JobSpec field, append to the appropriate list.
+- **`chart.imagePullSecretName`** — Resolves registry pull secret for pod templates. Precedence: per-job key with `hasKey` (`cronJob.jobs.*` / `job.jobs.*` only; `""` opts out) → truthy workload `imagePullSecrets` → `image.pullSecrets` → `image.imagePullSecrets` (alias, only when `image.pullSecrets` is unset). Deployment/StatefulSet `imagePullSecrets: ""` is falsy and inherits `image.pullSecrets`. Requires string secret names; guards `image` map access when `.Values.image` is null.
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `5c1b366`. patterns.md now reads "alias, only when `image.pullSecrets` is unset **or empty**", matching the implementation (`pullSecrets` defaults to `""`, and `""` is treated as equivalent to unset so the alias still applies) and the README/values.yaml wording. Doc-only, no behavior change; all 318 unit tests still pass.

templates/_helpers.tpl
```diff
@@ -162,6 +162,133 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+Resolve the Kubernetes imagePullSecrets secret name for a pod template context.
+Precedence (highest wins):
+  1. cronJob.jobs.<name>.imagePullSecrets / job.jobs.<name>.imagePullSecrets when set to a non-null value (only this path: empty string "" opts out; null is treated as unset and inherits from below). "" is the ONLY opt-out — other falsey values (false, 0, []) are rejected as type errors, not silently rendered as a secret name.
+  2. .imagePullSecrets on the merged workload context when truthy (deployment/statefulSet empty string is falsy and inherits steps 3-4)
+  3. .Root.Values.image.pullSecrets (when .Values.image is a map) — customer-requested top-level key
+  4. .Root.Values.image.imagePullSecrets (alias, only when image.pullSecrets is unset)
+Per-job lookup uses Values.jobs directly because shallow merge does not reliably override globals.
+Every level is kind-checked before the truthiness test that selects it, so a set-but-non-string
+value fails the render instead of being silently swallowed as "unset" and inheriting from below.
+*/}}
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `79f986e`. The `chart.imagePullSecretName` precedence comment (step 4) now says the `image.imagePullSecrets` alias applies "only when `image.pullSecrets` is unset **or empty**", so the comment matches the intentional behavior of treating `image.pullSecrets: ""` as equivalent to unset. Aligned with values.yaml and the patterns.md fix in `5c1b366`. Doc-only, no behavior change.

templates/_helpers.tpl
```diff
@@ -162,6 +162,133 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+{{- define "chart.jobPodGlobals" -}}
+{{- toYaml (pick . "imagePullSecrets") -}}
+{{- end -}}
+
+{{/*
+Whether a cronJob.jobs.<name> or job.jobs.<name> entry should render. .enabled must be boolean when set.
+A null .enabled is treated as unset (renders), matching how a null imagePullSecrets is treated as
+unset by chart.imagePullSecretName — this keeps `--set jobs.<name>.enabled=null` (Helm's canonical
+unset idiom) working rather than failing the render.
```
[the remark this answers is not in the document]
yesterdaysrebel: Valid — fixed in `44df35f`. The example now uses the real values paths (`--set cronJob.jobs.<name>.enabled=null` / `--set job.jobs.<name>.enabled=null`); there is no top-level `jobs` key. Doc-only, no behavior change; 318 unit tests still pass.

templates/job.yaml
```diff
@@ -1,5 +1,6 @@
-{{ if .Values.job.enabled }}
+{{ if and .Values.job .Values.job.enabled }}
```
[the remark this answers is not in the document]
yesterdaysrebel: Fixed in `47c2df4` — job.yaml:1 now uses `{{- if ... }}`, matching the convention in deployment.yaml/statefulset.yaml. One note: the final rendered manifest is byte-identical before/after (helm's manifest splitter already strips the leading blank line, verified with a full `helm template` diff), so this is a pure style/consistency change rather than a rendering fix. 318 tests pass, lint clean.

templates/cronjob.yaml
```diff
@@ -1,5 +1,6 @@
-{{ if .Values.cronJob.enabled }}
+{{ if and .Values.cronJob .Values.cronJob.enabled }}
```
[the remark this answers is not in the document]
yesterdaysrebel: Fixed in `47c2df4` — cronjob.yaml:1 now uses `{{- if ... }}`, matching deployment.yaml/statefulset.yaml. Same caveat as the job.yaml thread: verified the full rendered manifest is byte-identical before/after (the leading blank line is already stripped by helm), so this is style consistency, not a behavior change.

templates/cronjob.yaml
yesterdaysrebel: fixed

templates/_helpers.tpl
yesterdaysrebel: fixed

templates/_helpers.tpl
yesterdaysrebel: fixed

values.yaml
yesterdaysrebel: fixed

values.yaml
yesterdaysrebel: fixed

templates/_helpers.tpl
yesterdaysrebel: fixed

Chart.yaml
yesterdaysrebel: fixed

templates/_helpers.tpl
```diff
@@ -162,6 +162,134 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+{{- $explicitPerJob := false -}}
+{{- if and .name (or (eq .resourceType "cronJob") (eq .resourceType "job")) -}}
+{{- $jobsMap := dict -}}
+{{- if eq .resourceType "cronJob" -}}
+{{- $jobsMap = .Root.Values.cronJob.jobs -}}
+{{- else -}}
+{{- $jobsMap = .Root.Values.job.jobs -}}
+{{- end -}}
+{{- $jobPath := printf "%s.jobs.%s.imagePullSecrets" .resourceType .name -}}
+{{- with index $jobsMap .name -}}
```
venkatamutyala, line 244: This is the gap I'd want closed before merge.

Swapped this to `index $jobsMap (keys $jobsMap | sortAlpha | first)` as a mutation test — all 318 tests still pass, but every job silently takes the first job's secret:

```
real:   aaa -> secret-a   bbb -> secret-b   ccc -> TOP-CRED
mutant: aaa -> secret-a   bbb -> secret-a   ccc -> secret-a
```

No test has two jobs with different `imagePullSecrets` — the existing multi-job tests only cover labels/backoffLimit/enabled. Since `index $jobsMap .name` sits inside a `range`, this is exactly where cross-job leakage would happen, and several differently-named cronjobs is the whole use case here. A 3-job test with disagreeing secrets kills it.

values.yaml
```diff
@@ -67,6 +67,21 @@ image:
...
+  # NOTE: the top-level key here is `pullSecrets`, but every workload/per-job override key is
+  #   `imagePullSecrets` (not `pullSecrets`). Using `pullSecrets` on a workload is a silent no-op.
+  # Only cronJob.jobs.<name>.imagePullSecrets / job.jobs.<name>.imagePullSecrets: "" opts out (no pull
+  #   secret); an empty string elsewhere inherits, and a `null` per-job value behaves as unset.
+  # At this top level "" is equivalent to unset: it does not block the `imagePullSecrets` alias below.
+  #   To render no pull secret at all, leave both keys empty.
+  # Example: pullSecrets: my-registry-cred
+  pullSecrets: ""
+  # -- Alias for `image.pullSecrets`, honored only when `pullSecrets` is unset or empty.
+  # Prefer `pullSecrets`; this exists because both spellings were acceptable to the requester.
```
venkatamutyala: Can we cut the alias? Two spellings for one thing means a precedence rule ("honored only when pullSecrets is unset or empty") that nobody will remember, and it's permanent API once this tags. `pullSecrets` on its own is fine.

values.yaml
```diff
@@ -83,6 +98,9 @@ commonLabels:
 deployment:
   # -- Whether to create a Deployment
   enabled: false
+  # -- Registry pull secret for this Deployment (overrides `image.pullSecrets`). Empty string inherits `image.pullSecrets`.
```
venkatamutyala: There's no way to opt a Deployment out of a global — both `""` and `null` inherit it:

```
image.pullSecrets: G  +  deployment.imagePullSecrets: ""    -> G
image.pullSecrets: G  +  deployment.imagePullSecrets: null  -> G
```

Meanwhile `cronJob.jobs.<name>.imagePullSecrets: ""` opts out. Same literal, opposite meaning depending on level. So once a common values file sets a secret, every Deployment/StatefulSet in the fleet is stuck with it while jobs get an escape hatch.

templates/_helpers.tpl
```diff
@@ -162,6 +162,134 @@ Object fields: structured YAML (maps/lists). Add new K8s fields to $objectFields
...
+     literal secret name ("false", "0", "[]"). */}}
+{{- if not (and (kindIs "string" $secret) (eq $secret "")) -}}
+{{- if not (kindIs "string" $secret) -}}
+{{- fail (printf "%s must be a string secret name, got %s (value: %v)" $source (kindOf $secret) $secret) -}}
+{{- end -}}
+{{- $secret = trim $secret -}}
+{{- if eq $secret "" -}}
+{{- fail (printf "%s must not be empty or whitespace only" $source) -}}
+{{- end -}}
+{{- if regexMatch "\\s" $secret -}}
```
venkatamutyala, line 283: Minor: this guard has no test. Disabled it and 318/318 still pass, and `image.pullSecrets: "my secret"` happily renders `- name: "my secret"`.

Worth knowing it matters more than it looks — k8s doesn't validate pull secret names at all. Tested against a real API server and it accepts `My_Secret`, `-bad`, `UPPER`, `foo/bar`, even names with spaces; the pod just runs and warns. So this render-time check is the only thing that ever catches a typo.

templates/_podTemplate.tpl
```diff
@@ -75,9 +75,10 @@ spec:
       topologyKey: topology.kubernetes.io/zone
       whenUnsatisfiable: ScheduleAnyway
   {{- end }}
-  {{- if .imagePullSecrets }}
+  {{- $imagePullSecret := include "chart.imagePullSecretName" . | trim }}
+  {{- if $imagePullSecret }}
   imagePullSecrets:
-  - name: {{ .imagePullSecrets }}
+  - name: {{ $imagePullSecret | quote }}
```
venkatamutyala, line 85: Worth a short comment on why the quote is here. helm-unittest can't catch its removal (it parses with a YAML 1.2 parser and normalizes the difference away), but Helm's actual loader is YAML 1.1 — so an unquoted `y` or `no` becomes a bool. Easy for someone to "simplify" this away later with a green suite.

venkatamutyala (changes requested):
Feature works — checked that `image.pullSecrets` actually reaches cronjob pods, and the phantom-cronjob-from-common-values problem is gone. Rendered all the examples plus some realistic tenant values against 0.13.0 and got no diffs beyond the version label and secret-name quoting, so the compat story holds up.

A few things before this goes in:

- Multi-job isolation isn't covered by any test (left a comment inline). That's the one that actually worries me.
- Drop the `image.imagePullSecrets` alias — John's "or ... if you prefer that naming" reads as pick one, not ship both. Free to drop now, permanent once 0.14.0 is tagged.
- `""` means opposite things depending on level, and there's no way to opt a Deployment/StatefulSet out of a global at all (inline).
- Could you paste a fleet-wide grep for `enabled:` under `cronJob.jobs` / `job.jobs`? If any tenant has it set today it's inert, so that cronjob is live right now — merging deletes it under prune with no values change. That's the only hard blocker for me.

Nit: body says 292 tests, it's 318 now.

Two pre-existing things I hit while reviewing — not this PR, worth their own issues:
- `statefulSet.enabled` + `cronJob.enabled` hangs `helm template` forever (same on main). `statefulset.yaml:33` passes `.Values.statefulSet` as the merge destination, so it gets a self-referential `Root` and the cronjob merge walks the cycle. Swapping the args to match `deployment.yaml:47` fixes it and tests stay green.
- `ct lint --chart-dirs=.` in `make lint` finds zero charts and exits 0, so nothing actually enforces the chart version bump today.

templates/_helpers.tpl
yesterdaysrebel: Valid, and reproduced — your mutant survived 318/318 here too. Fixed in `5fc4c61`.

Added a 3-job test with disagreeing secrets to both `cronjob_test.yaml` and `job_test.yaml`:

```
aaa -> secret-a   bbb -> secret-b   ccc -> top-cred (inherits image.pullSecrets)
```

The third job is deliberately secret-less so the test pins the fall-through too, not just the two explicit values. Also added a companion test that a per-job `""` opt-out stays scoped to that job while its sibling still inherits the workload global — same `range`, same leakage risk.

Re-ran your exact mutant against the new suite: `index $jobsMap (keys $jobsMap | sortAlpha | first)` now produces 3 failures instead of passing.

values.yaml
yesterdaysrebel: Agreed — cut in `5fc4c61`. `image.imagePullSecrets` is gone; `image.pullSecrets` is the only top-level key.

You were right that this was the expensive kind of convenience: the alias existed only because both spellings were acceptable to the requester, and it bought a precedence rule ("honored only when `pullSecrets` is unset **or empty**") that had already generated three separate review comments just to keep its own docs self-consistent. That is a bad trade for a key that tags as permanent API.

Dropped the helper branch, the values key, the alias assert, and the docs. Added a test asserting `image.imagePullSecrets` is now inert (a misspelling renders no secret) so nobody reintroduces it by accident. helm-docs regenerated — the README row is gone.

values.yaml
yesterdaysrebel: Valid — confirmed both cases resolved to `G` exactly as you show. Fixed in `5fc4c61`.

The same literal meaning opposite things depending on level was indefensible, so `""` now means one thing everywhere: **opt out**. Selection is `hasKey` + non-null at every level — a set value wins even when empty, `null`/absent inherits:

```
image.pullSecrets: G  +  deployment.imagePullSecrets: ""    -> no pull secret
image.pullSecrets: G  +  deployment.imagePullSecrets: null  -> G
```

This required changing the workload defaults from `imagePullSecrets: ""` to `null` in `values.yaml` — with a `""` default, the opt-out semantics would have opted the entire fleet out. `null` keeps the key visible to helm-docs while still meaning "unset".

Deployment/StatefulSet opt-out and null-inherit tests added on both paths; error hints now say "opt this workload out" rather than the old "inherit image.pullSecrets".

templates/_helpers.tpl
yesterdaysrebel: Valid — reproduced: deleting the guard leaves 318/318 green, and `image.pullSecrets: "my secret"` rendered `- name: "my secret"`. Fixed in `5fc4c61`.

The point about the API server not validating pull secret names is the part worth acting on, so I wrote it into the test block as the reason the guard exists — a future reader deciding whether this check earns its keep needs to know k8s will not catch it downstream. Coverage added for internal spaces, tabs, whitespace-only, the trim-to-valid case, and the per-job path (which names `cronJob.jobs.<name>.imagePullSecrets`).

Mutants now die: removing the whitespace guard fails 4 tests, removing the empty/whitespace-only guard fails 2.

templates/_podTemplate.tpl
yesterdaysrebel: Done in `5fc4c61` — comment added above the block.

Your reasoning is the comment: it records that Helm loads YAML 1.1 (so an unquoted `y`/`no`/`on`/`off` name becomes a bool) **and** that helm-unittest cannot catch the removal because it reads the rendered doc with a YAML 1.2 parser. The second half is the part that matters — without it the note reads as trivia, and the trap is precisely that a green suite looks like permission to simplify it away.

One note: my first attempt placed the comment between `imagePullSecrets:` and `- name:`, which injected a stray blank line into the rendered PodSpec. Caught it on a real `helm template` (the unit tests passed either way — same YAML-parser blind spot). Moved above the `if` and re-verified the render is byte-identical.

venkatamutyala (approved):
## LGTM 🚀✅🎉

🙌🔥💯 Nicely done — this addresses every prior blocker and then some.

✅ **B1** version bumped to `0.14.0` + rebased onto main — no released-tag collision 🏷️
✅ **B2** `image.pullSecrets` (the customer's requested key) now works 🔑
✅ **M1** `ternary` nil-panic replaced with explicit `if/else` — no more render crash 💥➡️😌
✅ **M2** `chart.jobPodGlobals` narrowed to `imagePullSecrets` only — zero pod-spec leakage 🎯
✅ uniform `""` opt-out, loud type validation, `| quote` YAML-1.1 bool guard 🛡️

🧪 324/324 unit tests green · 🧹 `helm lint` clean · 🔁 backwards compatible (only cosmetic secret-name quoting vs `v0.13.0`, YAML-equivalent) · 🤝 both of John's asks solved (global secret inheritance + per-job `enabled: false` opt-out).

🧵 Non-blocking follow-ups: 🪤 the `image.pullSecrets` (top) vs `imagePullSecrets` (workload) key-name split is a documented silent-no-op footgun — a targeted `values.schema.json` (`not/required` guards) would make wrong keys fail loudly. 🏷️ heads-up on the pre-existing `v1.0.0` tag possibly sorting as latest, and 🔄 expect a fleet-wide pod roll on the version bump.

Ship it! 🚢😎👏

